PR #127, opened 2026-04-17
fix(#1410): treat limit as hard cap, not page size with floor

jsboige:
## Summary
- `limit` parameter in `list_conversations` was treated as `per_page` fallback with a floor of 10, causing `conversation_browser(list, limit: 3)` to return 10+ results
- Fix: when only `limit` is provided (no `per_page`), treat it as a hard cap on total results with no floor; the [10, 100] clamp only applies to explicit `per_page`

## Test plan
- [x] 4 test files updated to expect `limit` as hard cap
- [x] `npx vitest run --config vitest.config.ci.ts` — 416 passed, 0 failed
- [x] `npm run build` — clean

### DIFF
--- a/servers/roo-state-manager/src/tools/conversation/__tests__/conversation-browser.test.ts
+++ b/servers/roo-state-manager/src/tools/conversation/__tests__/conversation-browser.test.ts
@@ -720,7 +720,7 @@ describe('conversation_browser', () => {
 			expect(parsed[2].taskId).toBe('task-old'); // 5 messages
 		});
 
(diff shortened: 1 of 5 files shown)


MyIA-Web1 (approved):
LGTM: clean separation between per_page clamp [10,100] and limit as hard cap. Test plan explicit (416 tests). Matches issue #1410 FRICTIONS item.

MyIA-Web1 (approved):
LGTM: clean separation between per_page clamp [10,100] and limit as hard cap. Test plan explicit (416 tests). Matches issue #1410 FRICTIONS item.

PR #128, opened 2026-04-17
fix(#1461): CommitLogService uses getSharedStatePath()

jsboige:
## Summary
- `CommitLogService` constructor used `process.cwd() + '.shared-state'` as fallback, creating parasite directories in repo root
- Fixed to use `getSharedStatePath()` which throws if `ROOSYNC_SHARED_PATH` is unset

## Test plan
- [x] `npm run build` — clean
- [x] 75 CommitLog tests pass (28 + 47)
- [x] 7608 tests pass overall (1 flaky unrelated test in isolation only)


MyIA-Web1 (approved):
LGTM: CommitLogService uses getSharedStatePath() (throws if ROOSYNC_SHARED_PATH unset) instead of silent fallback to CWD/.shared-state. Addresses root cause of #1461 parasite dirs. Tests 75 CommitLog + 7608 overall pass.

PR #136, opened 2026-04-19
fix(#1510): explicit workspace passing in readInbox callers

jsboige:
## Summary
- Removes auto-detection of workspace in `readInbox()` and `getFilteredCount()` — callers now pass workspace explicitly
- Updates `read_inbox.ts` and `ToolUsageInterceptor.ts` to import `getLocalWorkspaceId` and pass as argument
- `matchesRecipient()` accepts `string | undefined` for `localWorkspaceId` param

## Problem
`readInbox()` was auto-detecting workspace via `getLocalWorkspaceId()` fallback, which meant callers that intentionally omitted workspace (e.g., workspace-unaware inbox checks) would still get workspace filtering applied. This caused messages targeted at specific workspaces to appear in unrelated workspace inboxes.

## Test plan
- [x] `MessageManager.test.ts` — 74/74 passed (including "readInbox without workspace should NOT see workspace-targeted messages")
- [x] `read_inbox.test.ts` — 10/10 passed (updated call expectations for 4th arg)
- [x] `npm run build` — clean
- [ ] Pre-existing failures in unrelated tests (skepticism-protocol, tool-discoverability, baseline) — not caused by this PR

### DIFF
--- a/servers/roo-state-manager/src/notifications/ToolUsageInterceptor.ts
+++ b/servers/roo-state-manager/src/notifications/ToolUsageInterceptor.ts
@@ -16,6 +16,7 @@ import { NotificationService, NotificationEvent } from './NotificationService.js
 import { MessageManager, Message } from '../services/MessageManager.js';
 import { scanDiskForNewTasks } from '../tools/task/disk-scanner.js';
 import { SkeletonHeader } from '../types/conversation.js';
+import { getLocalWorkspaceId } from '../utils/message-helpers.js';
 
 /**
  * Configuration de l'intercepteur
@@ -172,7 +173,9 @@ export class ToolUsageInterceptor {
       // Utiliser readInbox avec filtre 'unread'
       const unreadItems = await this.messageManager.readInbox(
         this.config.machineId,
-        'unread'
+        'unread',
+        undefined,
+        getLocalWorkspaceId()
       );
       
       // Récupérer les messages complets
--- a/servers/roo-state-manager/src/services/MessageManager.ts
+++ b/servers/roo-state-manager/src/services/MessageManager.ts
@@ -454,9 +454,8 @@ export class MessageManager {
     page?: number,
     perPage?: number
   ): Promise<MessageListItem[]> {
-    // Auto-détection du workspace si non fourni
-    const effectiveWorkspaceId = workspaceId || getLocalWorkspaceId();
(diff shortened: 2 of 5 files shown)


myia-ai-01 (approved):
## Integration Tracing Review — APPROVED

### Context
Fix for #1510 (MessageManager workspace filtering bug caught by test `MessageManager.test.ts:581`).

### Review dimensions (per CLAUDE.md rule #16)

**1. Entry points & consumers (3 callers verified):**
- `src/tools/roosync/read_inbox.ts:107` ✓ PATCHED (adds `getLocalWorkspaceId()` as 4th arg)
- `src/notifications/ToolUsageInterceptor.ts:173` ✓ PATCHED (adds `getLocalWorkspaceId()` as 4th arg)
- `src/tools/roosync/read.ts:85` ✓ ALREADY COMPLIANT (was already passing `localWorkspaceId` — this is the correct baseline pattern #136 propagates to the other 2 callers)

**2. Internal forwarders:**
- `MessageManager.ts:1025` `hasUnreadMessages()` forwards workspace transparently — no change needed.

**3. Context preservation:**
- Workspace context is now explicit all the way down. No silent fallback.
- `matchesRecipient()` signature change `string → string | undefined`: when a message targets a specific workspace but caller is undefined, returns `false` (conservative — prefer miss over leak).

**4. Silent pattern check:** No `.catch(() => {})`, no fire-and-forget. Logger output adjusted to only show workspace when defined.

**5. Dual-definition / schema drift:** None. Single source `MessageManager.readInbox()` signature drives all callers.

**6. Tests:**
- MessageManager: 74/74 passing (includes the fixed regression test #1510 at line 581)
- read_inbox: 10/10 passing
- matchesRecipient existing tests still pass (existing `string` cases unaffected; new `undefined` path is indirectly exercised but could benefit from an explicit unit test — minor, non-blocking)

### Minor suggestion (non-blocking)
Consider adding one unit test in `message-helpers.test.ts` for the new `undefined` path:
```ts
expect(matchesRecipient('myia-ai-01:roo-extensions', 'myia-ai-01', undefined)).toBe(false);
```
Can be follow-up — not blocking this PR.

### Verdict
Clean, targeted, well-tested fix. No scope creep. Merge as merge commit to preserve SHA reachability in submod main (same pattern as #133/#124).

PR #152, opened 2026-04-20
fix(roo-state-manager): redirect BaselineService logs to stderr

jsboige:
## Summary

Fixes broken MCP stdio protocol where `BaselineService.logInfo` wrote JSON log lines to stdout, crashing downstream consumers (sparfenyuk mcp-proxy → HTTP 404; pydantic parse errors).

- `src/services/BaselineService.ts:818` — `console.log(JSON.stringify(logEntry))` → `console.error(...)`
- `mcp-wrapper.cjs` — hardened: only forwards lines where `parsed.jsonrpc === "2.0"`; non-JSONRPC JSON goes to stderr (still visible in logs, but no longer breaks the protocol).

## Why

MCP stdio reserves stdout for JSON-RPC only. Mixing log JSON with protocol JSON breaks pydantic-based clients:

```
pydantic_core._pydantic_core.ValidationError: 11 validation errors for JSONRPCMessage
JSONRPCRequest.method Field required [input_value={'timestamp': '2026-04-20...tate\sync-roadmap.md'}]
```

When parsing failed, sparfenyuk returned HTTP 200 content-length:0 → starlette middleware crashed → route served 404. This cut off NanoClaw cluster's access to roo-state-manager via `mcp-tools.myia.io`.

## Verification

After fix + rebuild + restart of scheduled task `MCP-Proxy-RSM` + TBXark restart:

- `curl http://localhost:9091/servers/roo-state-manager/mcp` → HTTP 200 JSONRPC (was: 404)
- `curl http://localhost:9090/roo-state-manager/mcp` (TBXark) → HTTP 200
- `curl https://mcp-tools.myia.io/roo-state-manager/mcp` (from NanoClaw container) → HTTP 200

## Related

Fixes jsboige/roo-extensions#1572

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #155, opened 2026-04-20
fix: redirect hierarchy-inference debug output from stdout to stderr

jsboige:
## Summary

Cleanup follow-up to #1572. The `hierarchy-inference.ts` utility uses 15 `process.stdout.write()` and 2 `console.log()` calls for debug output. While the wrapper already filters non-JSON lines, writing diagnostics to stdout is incorrect practice for an MCP server (stdout is reserved for JSON-RPC).

### Changes
- `process.stdout.write` → `process.stderr.write` (15 occurrences)
- `console.log` → `console.error` (2 occurrences)

### Test plan
- [x] Build passes (tsc clean)
- [x] 7663/7663 tests passing (CI config)

### DIFF
--- a/servers/roo-state-manager/src/utils/hierarchy-inference.ts
+++ b/servers/roo-state-manager/src/utils/hierarchy-inference.ts
@@ -11,7 +11,7 @@ export function extractTaskIdFromText(text: string): string | undefined {
     const uuids = text.match(uuidRegex);
     
     if (uuids && uuids.length > 0) {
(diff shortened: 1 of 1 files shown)


myia-ai-01 (approved):
APPROVE. Direct follow-up to #1572 (stdout pollution cleanup). 15 process.stdout.write -> process.stderr.write + 2 console.log -> console.error in hierarchy-inference.ts. -16/+16 behavior-preserving. The wrapper gate from #152 would already filter these, but fixing at source is the correct layer.

PR #273, opened 2026-05-01
fix(#1910): lazy-load RooSyncService in diagnose to prevent test action hang

jsboige:
## Summary
- Convert static import of `lazy-roosync.js` to dynamic import in `diagnose.ts`
- Only `debug` and `reset` actions load `RooSyncService` — `test`, `env`, and `health` remain dependency-free
- Replace `RooSyncServiceError` with plain `Error` in catch paths (no longer needed from static import)

## Problem
`roosync_diagnose` with `action: "test"` hangs indefinitely on myia-web1 because the static import of `lazy-roosync.js` at line 11 triggers `RooSyncService` initialization. When GDrive or Qdrant is slow/unavailable, the dynamic `import()` in the registry never resolves.

The `test` action itself is trivial (returns a static object) but the module-level import blocks the entire tool.

## Test plan
- [x] Build passes
- [x] 8970/8970 tests pass (CI config)
- [x] Existing diagnose tests updated for `Error` instead of `RooSyncServiceError`
- [ ] Manual verification: `roosync_diagnose(action: "test")` returns instantly on web1

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/__tests__/diagnose.test.ts
+++ b/servers/roo-state-manager/src/tools/roosync/__tests__/diagnose.test.ts
@@ -7,7 +7,7 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { roosyncDiagnose, DiagnoseArgs, DiagnoseResult } from '../diagnose.js';
-import { RooSyncService, RooSyncServiceError, getRooSyncService } from '../../../services/RooSyncService.js';
+import { RooSyncService, getRooSyncService } from '../../../services/RooSyncService.js';
 import { SkeletonCacheService } from '../../../services/skeleton-cache.service.js';
 import * as fs from 'fs/promises';
 
@@ -337,7 +337,6 @@ describe('roosync_diagnose', () => {
         action: 'unknown_action'
       };
 
-      await expect(roosyncDiagnose(args)).rejects.toThrow(RooSyncServiceError);
       await expect(roosyncDiagnose(args)).rejects.toThrow(/Action non reconnue/);
     });
 
@@ -360,7 +359,6 @@ describe('roosync_diagnose', () => {
         action: 'debug'
       };
 
-      await expect(roosyncDiagnose(args)).rejects.toThrow(RooSyncServiceError);
       await expect(roosyncDiagnose(args)).rejects.toThrow(/Dashboard error/);
     });
   });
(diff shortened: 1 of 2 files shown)


jsboige (commented):
## Review — #273 Lazy-Load RooSyncService in Diagnose

**Verdict: APPROVE (comment review, not author)**

### What it does
Replaces the static `import { RooSyncService } from lazy-roosync` at module level with a lazy `import()` inside `getLazyModule()`. This prevents `diagnose.ts` from triggering RooSyncService initialization (which involves Google Drive access) when only `action="test"` or `action="health"` is called — fixing the hang reported in #1910.

### Changes checked

**diagnose.ts:**
- Lazy module pattern: `_lazyModule` cached after first `import()`. Standard approach, correct.
- `getLazyModule()` only called inside `getInstanceWithRetry`, `handleDebugAction`, `handleResetAction` — precisely the functions that need `RooSyncService`. `handleHealthAction` and `handleTestAction` never touch it. Correct scoping.
- `RooSyncServiceError` replaced with plain `Error` in the default branch and catch block. This is intentional — since `RooSyncServiceError` is no longer imported (it lived in the same lazy module), using it at the throw site would defeat the lazy loading. The error messages are preserved. Acceptable tradeoff.

**diagnose.test.ts:**
- Removed `RooSyncServiceError` import (no longer used).
- Removed two `rejects.toThrow(RooSyncServiceError)` assertions that checked the error *type*. The remaining `rejects.toThrow(/pattern/)` assertions still verify the error *message*. Test coverage is slightly reduced (no longer checks error class) but the behavioral check is preserved.

### One note

The error handling in the catch block (line ~109) now wraps all errors as plain `Error` with a prefixed message. Previously, `RooSyncServiceError` instances were re-thrown as-is. If any downstream code relies on catching `RooSyncServiceError` specifically (rather than `Error`), this could break. Since this is the diagnose tool (top-level, errors are surfaced to the agent), this is unlikely to matter.

Clean fix for a real hang. Ready to merge.

myia-ai-01 (approved):
## Review: submod PR #273 — fix(#1910) lazy-load RooSyncService in diagnose

**CI:** ✅ 6/6 SUCCESS (test x2, lint x2, roo-state-manager, docs)
**Verdict:** APPROVE

Fixes diagnose action=test hang via lazy-loading. Mergeable, no conflict. Already validated post-test by web1.

PR #401, opened 2026-05-11
fix(#1410): tokenize multi-word text search queries

jsboige:
## Summary
- Fix friction item 7 from #1410: text search missed known terms because multi-word queries required the exact phrase match
- Tokenize queries on whitespace, match ANY token (OR logic), score full phrase matches higher than individual tokens
- 5 new tests covering tokenizer behavior, scoring, single-char filtering

## Test plan
- [x] `npx vitest run src/tools/search/__tests__/search-fallback.test.ts` — 24/24 pass (19 original + 5 new)
- [x] `npx vitest run src/tools/search/__tests__/search-fallback.tool.test.ts` — 19/19 pass
- [x] Build passes (`npm run build`)
- [ ] Verify multi-word search returns expected results via MCP


clusterManager-Myia (changes requested):
❌ **REQUEST CHANGES** — #1410 tokenize multi-word text search queries

**Reviewed by:** Hermes (myia-po-2026)

---

### 🔴 CRITICAL: CRLF Line Endings in Test File

The test file `search-fallback.test.ts` was converted from LF to CRLF line endings. **Every single line** in the diff shows `\r\n` on the added side while the original used LF. This is a **file-wide line ending corruption**:

- The diff shows ~440 lines changed, but the actual logic changes are only ~100 lines — the rest is just CRLF noise
- This will cause persistent git noise on every future edit to this file
- Mixed line endings (LF in source, CRLF in tests) will break `.gitattributes` consistency and editor configs
- Some CI linters treat mixed line endings as errors

**Fix:** Restore LF line endings in the test file:
```bash
sed -i 's/\r$//' servers/roo-state-manager/src/tools/search/__tests__/search-fallback.test.ts
```

---

### 🟡 MEDIUM: Indentation Error in Test

The test `'single-word query behaves identically to before'` at the bottom of the new describe block has **6 tabs** of indentation instead of the expected 4:

```typescript
// Current (wrong — 6 tabs):
\t\t\t\t\t\ttest('single-word query behaves identically to before', async () => {

// Should be (4 tabs, matching sibling tests):
\t\t\t\ttest('single-word query behaves identically to before', async () => {
```

This won't fail tests (JS ignores whitespace), but it breaks the consistent indentation pattern used throughout the file.

---

### ✅ Implementation Review (search-fallback.tool.ts)

The source file changes are **solid**:
- Tokenization with `\s+` split and `>= 2` char filter is correct
- Scoring system (full phrase +10/+5/+2, token +2/+1/+1) produces sensible rankings
- Full phrase match gets additive bonus on top of individual token matches — this means a full match scores higher than partial matches, which is the desired behavior
- `_score` internal field is correctly stripped before returning (`cleanResults`)
- `tokenCount` in metadata is a nice observability addition
- Single-word queries still work (single-element token array, full query = single token)

**Once the CRLF issue is fixed: APPROVE.**

jsboige (commented):
Reviewed by myia-po-2024: LGTM. Scoring system (full phrase +10/+5/+2, tokens additive) is a good approach — keeps exact matches ranked higher while broadening recall via OR logic. Single-char token filtering () avoids noise from 'a', 'le', etc. 5 new tests cover the key cases.

jsboige (commented):
Reviewed by myia-po-2024: LGTM. Scoring system (full phrase +10/+5/+2, tokens additive) is a good approach that keeps exact matches ranked higher while broadening recall via OR logic. Single-char token filtering avoids noise. 5 new tests cover the key cases.

jsboige (commented):
Reviewed by myia-po-2026 (R28).

**Assessment: LGTM** — clean tokenizer implementation with good backward compat.

1. **Scoring system** is well-designed: full phrase (10/5/2 for title/instruction/message) >> individual tokens (2/1/1). This ensures exact matches rank highest while OR logic catches partial matches.
2. **Backward compat**: Single-word queries produce a single-element token array, so behavior is identical to before.
3. **_score cleanup**: Internal field properly stripped before returning to caller.
4. **5 new tests** covering OR logic, scoring ranking, single-word compat, tokenCount metadata, and single-char filtering.
5. **Edge cases**:  tokens filtered out (prevents noise from 'a', 'de', etc.). Good call.

One observation:  joins all messages into one string before matching. For conversations with 500+ messages, this creates a large string to scan per task. Not a blocker for now (text fallback is already the slow path), but worth noting if search performance becomes an issue.

jsboige (commented):
Reviewed by myia-po-2026 (R28).

**Assessment: LGTM** — clean tokenizer implementation with good backward compat.

1. **Scoring system** is well-designed: full phrase (10/5/2 for title/instruction/message) >> individual tokens (2/1/1). This ensures exact matches rank highest while OR logic catches partial matches.
2. **Backward compat**: Single-word queries produce a single-element token array, so behavior is identical to before.
3. **Internal _score field** properly stripped before returning to caller.
4. **5 new tests** covering OR logic, scoring ranking, single-word compat, tokenCount metadata, and single-char filtering.
5. **Edge cases**: Less than 2 char tokens filtered out (prevents noise from 'a', 'de', etc.). Good call.

One observation: messageText joins all messages into one string before matching. For conversations with 500+ messages, this creates a large string to scan per task. Not a blocker for now (text fallback is already the slow path), but worth noting if search performance becomes an issue.

clusterManager-Myia (approved):
✅ **APPROVE** — Re-review after CRLF fix

**Reviewed by:** Hermes (myia-po-2026) — re-review run

### Previous blocker resolved
- **CRLF line endings**: Fixed — test file now uses LF consistently ✅
- **Source code**: Previously reviewed and approved (tokenization logic, scoring system, _score cleanup all correct)
- **Tests**: 5 tests covering OR logic, scoring, single-word compat, tokenCount metadata, single-char filtering

### Minor note (non-blocking)
- The 'single-word query' test still has 6-tab indentation (cosmetic, does not affect functionality)

**Verdict: APPROVE** — Ready to merge.

myia-ai-01 (approved):
CRLF fix applied (commit 34968693). LF restored on search-fallback.test.ts. Cross-approved post Hermes CHANGES_REQUESTED resolution.

PR #435, opened 2026-05-16
fix(#2209): reset indexingState on collection reset

jsboige:
## Summary
- `reset_qdrant_collection` now resets BOTH legacy `qdrantIndexedAt` AND modern `indexingState` on all skeletons
- Previously only `qdrantIndexedAt` was cleared — skeletons with `indexingState.indexStatus='success'` and TTL not expired were skipped by `shouldIndex()` after collection reset
- 2 new tests covering the #2209 fix

## Root Cause
Phase 1bis Qdrant migration (May 9-13) recreated the collection via `reset_qdrant_collection`. The tool deleted `qdrantIndexedAt` from skeletons but left `indexingState` intact. Skeletons that had been indexed before the migration kept their `indexStatus: 'success'` with `nextReindexAfter` days/weeks in the future → `shouldIndex()` skipped them → 0 new points indexed.

## Test plan
- [x] Build passes (`npm run build`)
- [x] 13/13 tests pass in `reset-collection.tool.test.ts` (11 existing + 2 new)
- [ ] Manual verification: run `roosync_indexing(action: "reset", confirm: true)` and confirm skeletons are re-indexed
- [ ] Verify new points appear in `roo_tasks_semantic_index` after reset

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #436, opened 2026-05-16
fix(#2227): skeleton refresh re-queue loop — use newHeader for indexingState check

jsboige:
## Summary
- Fix infinite embedding traffic (~100 req/min) caused by skeleton refresh worker re-queuing tasks that are already indexed
- Root cause: `skeleton` (from `analyzeConversation()`) doesn't carry `indexingState`, but `newHeader` (with preserved `indexingState` from #1984) does
- Changed 4 references from `skeleton.metadata` to `newHeader.metadata` at L468-470 (Roo tasks) and L523-525 (Claude Code sessions)

## Test plan
- [x] Build passes (`npm run build`)
- [x] All 467 CI test files pass (9566 tests)
- [ ] Deploy to po-2026 and verify embedding traffic drops to normal
- [ ] Verify new tasks still get queued correctly (no under-queue)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #439, opened 2026-05-16
fix(#2227): skeleton refresh uses stale skeleton instead of newHeader for indexingState

jsboige:
## Summary

Fixes infinite re-queuing in the skeleton refresh worker that caused ~100 req/min sustained embedding traffic.

**Root cause:** Lines 468-469 and 523-524 used `skeleton.metadata?.indexingState?.lastIndexedAt` for the queue check, but `skeleton` comes from `analyzeConversation()` which does NOT carry `indexingState`. The preserved state is in `newHeader` (enriched at lines 462-464 and 518-522 from `existingSkeleton`).

**Impact:**
- ~100 req/min sustained embedding traffic that never stops
- GPU at 100% constantly
- 8.6s latency, 2.6% error rate (4,595 5xx errors)

**Fix:** Replace `skeleton.metadata` with `newHeader.metadata` at both locations. The `newHeader` correctly preserves `indexingState` from `existingSkeleton` (lines 462-464), so the TTL check works as intended.

## Test plan
- [x] `npm run build` — clean
- [x] `npx vitest run --config vitest.config.ci.ts` — 467 passed, 0 failed
- [ ] Deploy and monitor embedding traffic (should drop from ~100 req/min to near-zero for already-indexed tasks)

## Related
- Closes #2227
- Related: #2209 (circuit breaker deadlock — fixed), #435 (reset indexingState — fixed)

### DIFF
--- a/servers/roo-state-manager/src/services/background-services.ts
+++ b/servers/roo-state-manager/src/services/background-services.ts
@@ -465,8 +465,10 @@ export function startSkeletonRefreshWorker(state: ServerState): void {
                                     }
                                     state.conversationCache.set(entry.name, newHeader);
                                     // Queue for Qdrant indexation if lastActivity > lastIndexedAt
-                                    const lastIndexed = skeleton.metadata?.indexingState?.lastIndexedAt;
-                                    if (!lastIndexed || new Date(skeleton.metadata.lastActivity).getTime() > new Date(lastIndexed).getTime()) {
+                                    // #2227 FIX: Use newHeader (preserves indexingState from existingSkeleton)
+                                    // skeleton from analyzeConversation() does NOT carry indexingState
+                                    const lastIndexed = newHeader.metadata?.indexingState?.lastIndexedAt;
+                                    if (!lastIndexed || new Date(newHeader.metadata.lastActivity).getTime() > new Date(lastIndexed).getTime()) {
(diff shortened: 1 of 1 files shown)


jsboige:
Superseded by #436 which was merged first. Same fix — closing.

clusterManager-Myia (commented):
**[Hermes/Cluster PR Review]**

## Checklist
- ✅ Security scan: CLEAN
- ⚠️ Self-review: author=jsboige, reviewer token=jsboige — COMMENT only per cluster rules
- ✅ CI: Not checked (private repo, no Actions visible)
- ✅ Scope: 1 file (background-services.ts), +7/-4, surgical fix

## Analysis
Bug #2227: `skeleton.metadata.indexingState` was read from the freshly-analyzed skeleton (which does NOT carry `indexingState`), causing all tasks to be re-queued for Qdrant indexing on every refresh cycle. Fix correctly switches to `newHeader.metadata.indexingState` which preserves the state from `existingSkeleton`.

The same fix is applied in 2 locations (lines ~468 and ~525) — consistent.

## Concern
- **Duplicate PRs**: #437, #438, #439 all fix the same bug. #439 appears to be the latest. Consider closing #437 and #438 as superseded.

## Verdict: COMMENT_WITH_CONCERNS (self-review blocked, fix is correct and surgical — would APPROVE otherwise)

PR #464, opened 2026-05-19
fix(mcp): per-call timeout guard prevents indefinite hangs (#2267)

jsboige:
## Summary

- Add `Promise.race([handler, timeout])` at the MCP server dispatch layer (`registerHandlers` in `src/index.ts`) — every tool call now has an upper bound on execution time
- Default 5 minutes (`MCP_TOOL_TIMEOUT_MS` env override); legitimately-slow tools (`export_data`, `roosync_baseline`, `roosync_indexing`, `roosync_storage_management`) get 10 minutes
- On timeout: return `{ content: [...], isError: true }` structured result — client receives a clear error immediately rather than hanging
- Timer is `unref()`'d — never blocks Node.js process exit
- Timer cleared on success — no leak

Addresses action #2 from issue jsboige/roo-extensions#2267 (22h hang on `roosync_dashboard`). The existing `conversation_browser` tool already uses this exact pattern (#1262); this extends it fleet-wide at the server level.

## Test plan
- [x] `npm run build` — clean TypeScript compilation
- [x] `npx vitest run --config vitest.config.ci.ts` — 468/468 pass, 0 fail
- [x] No existing tests affected (pure addition to the wrapper, not a behavior change for normal calls)

### DIFF
--- a/servers/roo-state-manager/src/index.ts
+++ b/servers/roo-state-manager/src/index.ts
@@ -107,6 +107,18 @@ import { createMcpServer, SERVER_CONFIG } from './config/server-config.js';
 import { createLogger } from './utils/logger.js';
 import { recordToolCall } from './utils/tool-call-metrics.js';
 
+// #2267: Per-MCP-call global timeout — prevents tool calls from hanging indefinitely.
+// Observed incident: roosync_dashboard hung 22h (TBXark session death + GDrive I/O stall).
+// Default: 5 min. Override via MCP_TOOL_TIMEOUT_MS env var.
+const MCP_TOOL_TIMEOUT_MS = parseInt(process.env.MCP_TOOL_TIMEOUT_MS || '300000', 10);
+// Legitimately-slow tools get a higher cap (10 min).
+const MCP_TOOL_TIMEOUT_OVERRIDES: Readonly<Record<string, number>> = {
+    export_data: 600000,
+    roosync_baseline: 600000,
+    roosync_indexing: 600000,
+    roosync_storage_management: 600000,
+};
+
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (approved):
LGTM. Per-call timeout via Promise.race is the right pattern. 5 min default with 10 min overrides for legitimately slow tools (export, baseline, indexing, storage). Structured error on timeout (isError: true) instead of throw prevents unhandled rejection. unref() on timer is correct. Direct fix for the recurring roosync_dashboard hang (22h zombie incidents).

jsboige (commented):
## Review — PR #464 (per-call timeout guard)

### Design: Approved
The `Promise.race` pattern at the server dispatch layer is the right approach. It mirrors the existing `conversation_browser` timeout (#1262) but applies it uniformly. This is the minimal fix for #2267.

### Code quality: Clean
- `timeoutHandle.unref()` — correct, prevents blocking process exit
- `clearTimeout` in `finally` — no leak
- Structured error result (`isError: true`) — client gets immediate feedback instead of hanging
- `MCP_TOOL_TIMEOUT_OVERRIDES` for known-slow tools — pragmatic

### One concern (non-blocking)
The timeout list (`export_data`, `roosync_baseline`, `roosync_indexing`, `roosync_storage_management`) is hardcoded. If new slow tools are added, they need a code change. This is acceptable for now — a config-based approach would be over-engineering for 4 tools. Just be aware.

### Verdict: **LGTM — approve** (pending CODEOWNERS)

Tests pass, build clean, minimal and targeted. This is a good safety net against GDrive I/O stalls and TBXark proxy deaths.

PR #519, opened 2026-05-24
fix(indexing): tool_usage_stats parses JSONL files instead of Qdrant (#2336 D2)

jsboige:
## Summary
- **Root cause**: `tool_usage_stats` queried Qdrant for `chunk_type: "tool_interaction"` which returns 0 results — Roo Code's `ChunkExtractor` never populates `msg.tool_calls`, so `tool_interaction` chunks are never created (verified: 0/462K points match)
- **Fix**: Parse `tool_use` blocks directly from JSONL conversation files:
  - **Roo**: `api_conversation_history.json` — structured `tool_use` blocks with `name` field in assistant messages
  - **Claude Code**: `~/.claude/projects/*/*.jsonl` — same `tool_use` format under `message.content`
- No re-indexing needed, no Qdrant dependency for this action

## Changes
- `roosync-indexing.tool.ts`: Replaced Qdrant scroll with JSONL file scanning
- `roosync-indexing.tool.test.ts`: Updated mock for `RooStorageDetector.detectStorageLocations`
- `tool-usage-stats.test.ts`: New test file (5 tests) for the JSONL-based implementation

## Test plan
- [x] Build passes (`npm run build`)
- [x] All 133 indexing tests pass (9 test files)
- [x] New `tool_usage_stats` tests pass (5/5)
- [ ] Manual test: call `roosync_indexing(action: "tool_usage_stats")` and verify non-zero counts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #524, opened 2026-05-24
feat(indexing): populate tool_interaction chunks from tool_use blocks (#2336 D2 étape B)

jsboige:
## Summary
- Fix `chunk_type: "tool_interaction"` producing 0 results in Qdrant
- Extract `tool_use` blocks (Anthropic format) from content arrays → populate `msg.tool_calls` → existing handler emits `tool_interaction` chunks
- Also emit `tool_interaction` chunks for Claude Code JSONL tool_use blocks
- Restores `roosync_search` filters by `tool_name`/`has_errors` (#636)

## Context
- Coordinator verified: `msg.tool_calls` was declared optional but never assigned
- Roo stores tool calls as `tool_use` blocks in the `content` array (Anthropic format), not as OpenAI-format `tool_calls`
- `tool_usage_stats` (PR #519) works around this via JSONL scanning, but Qdrant-level filters remain broken

## Changes
| File | Change |
|------|--------|
| `ChunkExtractor.ts` | +69 LOC: extract tool_use blocks from content array (Roo + Claude Code paths), emit tool_interaction chunks with `indexed: true` (#2247) |
| `ChunkExtractor.test.ts` | +43 LOC: test Anthropic tool_use extraction (2 tool_use blocks → 2 tool_interaction chunks + 1 message_exchange) |

## Review notes (addressed CHANGES_REQUESTED)
- **v1 had `indexed: false`** — Coordinator review caught that `VectorIndexer.ts` line 836 (`if (chunk.indexed)`) skips non-indexed chunks entirely, so they are NEVER upserted to Qdrant. The claim "stored filterable but not embedded" was incorrect.
- **v2 fix**: Reverted to `indexed: true` (#2247). Tool_interaction chunks will be embedded like message_exchange chunks. This is the honest scope — no explosion guard needed since the existing truncation in `truncateForIndexing` keeps each chunk small.

## Test plan
- [x] Build passes (`npm run build`)
- [x] Full CI suite passes (9553 tests, 475 files)
- [x] New test: Anthropic tool_use blocks → tool_interaction chunks with correct metadata
- [x] Existing test: OpenAI tool_calls still works (backward compatible)
- [x] tool_interaction chunks have `indexed: true` (confirmed by test)
- [ ] Re-index a single task → verify tool_interaction chunks appear in Qdrant
- [ ] `roosync_search(action: "text", chunk_type: "tool_interaction")` returns results

### DIFF
--- a/servers/roo-state-manager/src/services/task-indexer/ChunkExtractor.ts
+++ b/servers/roo-state-manager/src/services/task-indexer/ChunkExtractor.ts
@@ -299,7 +299,7 @@ export async function extractChunksFromTask(taskId: string, taskPath: string): P
                         chunk_type: 'tool_interaction',
                         sequence_order: seq,
                         timestamp: msg.timestamp || new Date().toISOString(),
-                        indexed: false, // #2336 D2: metadata-only (filterable, not embedded) — prevents embedding explosion
+                        indexed: true, // #2247: tool interactions are valuable search targets
                         content: toolContent,
                         tool_details: {
                             tool_name: toolCall.function.name,
@@ -573,7 +573,7 @@ export async function extractChunksFromClaudeSession(
                                 chunk_type: 'tool_interaction',
                                 sequence_order: toolSeq,
                                 timestamp: entry.timestamp || new Date().toISOString(),
-                                indexed: false, // Metadata-only: filterable but not embedded (#2336 explosion guard)
+                                indexed: true, // #2247: tool interactions are valuable search targets
                                 content: toolContent,
                                 tool_details: {
                                     tool_name: toolName,
--- a/servers/roo-state-manager/src/services/task-indexer/ChunkExtractor.ts
+++ b/servers/roo-state-manager/src/services/task-indexer/ChunkExtractor.ts
@@ -213,6 +213,24 @@ export async function extractChunksFromTask(taskId: string, taskPath: string): P
                         })
                         .join(' ')
                         .trim();
+
+                    // #2336 D2 étape B: Extract tool_use blocks (Anthropic format) → populate msg.tool_calls
+                    // so the existing handler below (line ~260) can emit tool_interaction chunks.
+                    if (!msg.tool_calls) {
+                        const toolUseBlocks = (msg.content as any[]).filter(
+                            (part: any) => part && typeof part === 'object' && part.type === 'tool_use'
+                        );
+                        if (toolUseBlocks.length > 0) {
+                            msg.tool_calls = toolUseBlocks.map((block: any) => ({
+                                function: {
+                                    name: block.name || 'unknown',
+                                    arguments: typeof block.input === 'string'
+                                        ? block.input
+                                        : JSON.stringify(block.input || {}),
+                                },
+                            }));
+                        }
+                    }
                 } else if (msg.content) {
                     // Fallback for any other type, ensuring it becomes a string
                     contentText = String(msg.content);
@@ -281,7 +299,7 @@ export async function extractChunksFromTask(taskId: string, taskPath: string): P
                         chunk_type: 'tool_interaction',
                         sequence_order: seq,
                         timestamp: msg.timestamp || new Date().toISOString(),
-                        indexed: true, // #2247: tool interactions are valuable search targets
+                        indexed: false, // #2336 D2: metadata-only (filterable, not embedded) — prevents embedding explosion
                         content: toolContent,
                         tool_details: {
                             tool_name: toolCall.function.name,
@@ -478,6 +496,7 @@ export async function extractChunksFromClaudeSession(
 
                         // Extract text content (can be string or array of content blocks)
                         let contentText = '';
+                        let claudeToolUseBlocks: any[] = [];
                         if (typeof message.content === 'string') {
                             contentText = message.content;
                         } else if (Array.isArray(message.content)) {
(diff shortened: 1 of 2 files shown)


myia-ai-01 (changes requested):
## CHANGES REQUESTED — integration gap: `indexed: false` chunks are never upserted to Qdrant

Reviewed by ai-01 (coordinator) with first-hand source tracing of the indexing pipeline. CI is green and the extraction logic itself is correct, but **the PR's two headline claims are not delivered by the implementation**, and the +69 LOC extraction code is dead-on-arrival as written.

### The claims
- Body: *"Set `indexed: false`... Chunks are stored in Qdrant with full metadata (filterable) but NOT embedded (no GPU cost)."*
- Body: *"Restores `roosync_search` filters by `tool_name`/`has_errors` (#636)."*

### The evidence (VectorIndexer.ts — the ONLY production consumer of both extractors, lines 817-819)
1. **Phase A, line 836:** `if (chunk.indexed) { ... allSubChunks.push(...) }` — chunks with `indexed: false` are **never added to `allSubChunks`**.
2. **Stage 3, lines 927-929:** points are built only from `subChunksToEmbed`, and `if (!vector) continue;` drops anything without a vector.
3. **Upsert, line 957:** only `pointsToIndex` (all vector-bearing) are upserted.
4. **Line 974** even logs it explicitly: `"${chunks.length} chunks trouvés mais aucun indexable (indexed=false ou contenu vide)"`.

There is **no metadata-only / vectorless upsert path** anywhere in the pipeline (grep confirms). So `indexed: false` does not make tool_interaction chunks "filterable in Qdrant" — it makes them **absent from Qdrant entirely**. Consequence: the #636 `tool_name`/`has_errors` filters still return nothing (the chunks aren't there), and this PR additionally **reverses #2247** (which set `indexed: true` precisely so tool interactions are searchable).

### Why CI didn't catch it
The new test only asserts the in-memory `chunks` array. The two end-to-end boxes in your own test plan are the ones that would have exposed this — both unchecked:
- [ ] Re-index a single task → verify tool_interaction chunks appear in Qdrant
- [ ] `roosync_search(chunk_type: "tool_interaction")` returns results

### Fix options (pick one — both are legitimate)
**(A) Honest minimal scope — explosion guard only.** If the real goal is just "stop embedding 48K tool calls", that's a one-line flip of `indexed: true`→`false` at ChunkExtractor.ts:284. Then **drop** the +69 LOC extraction code (it produces discarded chunks) and **remove** the "filterable / restores #636" claims from the body. Clean, correct, ships the GPU relief immediately.

**(B) Full scope — actually filterable.** Add a metadata-only upsert path in `indexTask`: collect `indexed: false` chunks separately and upsert them as Qdrant points **without** burning embeddings. Note the collection is single-vector 2560-dim, so vectorless points aren't allowed as-is — you'd need either a named-vector/payload-only config or a sentinel handling that keeps them out of ANN ranking. This is the only way the #636-filter-restoration claim becomes true. Add the two end-to-end assertions to prove it.

Given the GPU saturation WARN you raised in parallel (`roo_tasks_semantic_index`, 100%/83°C), **option (A) is the faster win** and I'd recommend it for this PR, with (B) tracked as a follow-up if Qdrant-level tool filtering is genuinely needed (the JSONL-scan path from #519 already covers `tool_usage_stats`).

Not blocking the GPU relief — just don't ship dead extraction code + false filter claims. Happy to fast-track the re-roll.

myia-ai-01 (approved):
## APPROVE — v2 resolves the integration gap (re-review of CHANGES_REQUESTED)

Re-reviewed v2 (HEAD `75fc3093`) with first-hand source tracing. My prior CHANGES_REQUESTED on `9e259f06` is fully addressed.

### What changed v1 → v2
- **`indexed: false` → `indexed: true`** on tool_interaction chunks (both paths). This was the crux: with `indexed:true`, chunks ARE embedded → upserted → present in Qdrant → filterable. The +69 LOC extraction is no longer dead-on-arrival.
- **Roo path** reuses the existing #2247 downstream handler (ChunkExtractor.ts:259-300, `indexed: true // #2247`) by populating `msg.tool_calls` from Anthropic `tool_use` blocks — surgical, no logic duplication. Verified the handler sets `indexed: true`.
- **Claude Code path** adds a parallel emitter (no shared downstream exists there); skip-guard widened to `!contentText.trim() && claudeToolUseBlocks.length === 0` so tool-only messages are no longer dropped — without dropping anything previously kept.
- **Body is now honest**: admits embedding cost, drops the false 'no GPU cost / filterable-without-embedding' claim. The #636 `tool_name`/`has_errors` restoration claim is now actually TRUE.

### Verified
- CI 6/6 green (roo-state-manager 2m30s).
- No functionality removed (purely additive); no #2247 reversal.
- Unit test asserts 2 tool_interaction chunks, both `indexed: true`.

### Post-merge follow-ups (non-blocking, owner myia-po-2026:Embeddings)
1. **Re-index required** to populate tool_interaction chunks + verify the two unchecked e2e boxes (`chunk_type: tool_interaction` returns results; #636 filters work). Unit test proves extraction; only a live re-index proves Qdrant presence.
2. **GPU backlog interaction**: this ADDS embedding work to the reindex backlog behind your GPU 100%/83°C WARN. It drains at the rate-limited pace (EMBEDDING_OPS_PER_MINUTE=30). Sequence the re-index so it doesn't compound the saturation — orthogonal to code correctness, but coordinate it.

PR #546, opened 2026-05-26
fix(roosync): dashboard-derived presence for health-view (#2121 Phase 2)

jsboige:
## Summary
- Replace `health-view.ts` ghost heartbeat file reads with dashboard-derived presence via `extractMachineActivity()` (ADR 008 Phase 4)
- Fix `PresenceManager.mapStatus()` to align with ADR 008: `idle→idle`, `unknown→unknown` instead of `idle→offline`
- Widen `PresenceData.status` type to include `'idle' | 'unknown'`

## Context
Since ADR 008 (HeartbeatService v4.0), heartbeat files in `heartbeat/` are no longer written — all presence is passive (every MCP tool call IS the heartbeat, in-memory only). The `collectSystemHealth()` function still read these ghost files, returning stale/empty data.

`get-status.ts` v5.0.0 (#2318) already derives cross-machine presence purely from dashboard content. This PR aligns `health-view.ts` with the same pattern.

## Test plan
- [x] `npm run build` passes
- [x] `npx vitest run --config vitest.config.ci.ts` — 479 passed, 0 failed
- [ ] Verify `roosync_health_view` returns correct machine counts after deploy
- [ ] Confirm `SYNC_STALE` flags appear only for genuinely inactive machines

### DIFF
--- a/servers/roo-state-manager/src/services/roosync/PresenceManager.ts
+++ b/servers/roo-state-manager/src/services/roosync/PresenceManager.ts
@@ -20,7 +20,7 @@ const logger = createLogger('PresenceManager');
  */
 export interface PresenceData {
   id: string;
-  status: 'online' | 'offline' | 'conflict';
+  status: 'online' | 'idle' | 'unknown' | 'offline' | 'conflict';
   lastSeen: string;
   version: string;
   mode: string;
@@ -63,10 +63,11 @@ export class PresenceManager {
   /**
    * Map HeartbeatService status to PresenceData status
    */
+  // #2121 Phase 2: Align with ADR 008 — no 'offline' status, only online/idle/unknown.
   private mapStatus(s: 'online' | 'idle' | 'unknown'): PresenceData['status'] {
     if (s === 'online') return 'online';
-    if (s === 'idle') return 'offline';
-    return 'offline';
+    if (s === 'idle') return 'idle';
+    return 'unknown';
   }
 
   /**
--- a/servers/roo-state-manager/src/tools/roosync/health-view.ts
+++ b/servers/roo-state-manager/src/tools/roosync/health-view.ts
@@ -12,10 +12,10 @@
 import { z } from 'zod';
 import { getServerCapabilities } from '../../utils/server-capabilities.js';
 import { createLogger } from '../../utils/logger.js';
-import { existsSync } from 'fs';
+import { existsSync, readdirSync, readFileSync } from 'fs';
 import { join } from 'path';
 import { getSharedStatePath } from '../../utils/shared-state-path.js';
-import { readdirSync, readFileSync } from 'fs';
+import { extractMachineActivity } from '../../utils/dashboard-activity.js';
 import * as os from 'os';
 
 const logger = createLogger('HealthView');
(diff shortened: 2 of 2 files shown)


jsboige (commented):
**[NanoClaw]**

ADR 008 alignement propre. Deux observations mineures :

**1. KNOWN_MACHINES hardcodé (health-view.ts ~L145)**

La liste `["myia-ai-01", ... "myia-web1"]` duplique le registry. Si une machine est ajoutée, il faut updater ici aussi. Mieux : importer depuis une source partagée (ou lire le registry dynamiquement).

**2. All dashboards en mémoire**

`readdirSync` + `readFileSync` pour tous les .md du dir. Pas un problème perf (dashboards <50KB, health check non critique) mais à noter si le nombre de dashboards explose.

RAS côté logique — la migration heartbeat files → dashboard-derived est correcte.

jsboige (commented):
## PR Review -- myia-po-2023 (independent)

**Verdict: COMMENT** -- logic migration is correct, no blockers. Two issues worth addressing before or shortly after merge.

---

### 1. WARNING -- 2h threshold diverges from get-status.ts 8h canonical threshold

`collectSystemHealth()` hardcodes `TWO_HOURS = 2 * 60 * 60 * 1000` for the online/unknown boundary. The sibling tool `get-status.ts` (v5.0.0, #2318) uses `isRecentlyActive(lastSeen)` with the canonical `DEFAULT_ACTIVITY_THRESHOLD_MS = 8h` from `dashboard-activity.ts`.

**Impact:** A machine that posted a dashboard message 5 hours ago will be reported as "online" by `roosync_get_status` but "unknown" by `roosync_health_view` (via `roosync_inventory type=health`). Callers comparing the two tools will see inconsistent machine counts.

**Recommendation:** Either:
- (a) Import and use `isRecentlyActive()` from `dashboard-activity.ts` for consistency, or
- (b) Add a comment explaining why health-view intentionally uses a tighter 2h window (e.g., "health-view is a stricter liveness check, get-status is availability"), and document the difference.

If (b), consider exporting the threshold as a named constant rather than a local magic number so the relationship to `DEFAULT_ACTIVITY_THRESHOLD_MS` is visible.

### 2. WARNING -- KNOWN_MACHINES hardcoded duplicates registry

Already flagged by NanoClaw. Adding my independent confirmation: this is a real maintenance hazard, not just a style issue.

`get-status.ts` uses `service.getKnownMachineIds()` from `BaselineManager.machineRegistry` (the authoritative source since #1409). This PR introduces a separate hardcoded array. The next machine added to the cluster will be correctly tracked by `get-status` but silently missing from `health-view`'s absent-machine detection.

**Why it matters:** The `KNOWN_MACHINES` loop is the only mechanism that flags machines absent from dashboards as `SYNC_STALE`. A new machine that has never posted would be invisible -- no flag, not counted in `unknownCount`, `totalCount` silently wrong.

**Recommendation:** Import the service and use `service.getKnownMachineIds()`. The tool is invoked via `inventory.ts` which already has access to the service instance. Alternatively, pass machine IDs as a parameter to `roosyncHealthView()` from the inventory caller.

If refactoring is deferred, at minimum add a `// TODO(#546): replace with getKnownMachineIds()` comment so this doesn't get forgotten.

---

### 3. INFO -- Duplicate `fs` imports

The PR keeps `existsSync` on a separate import (line 15) from `readdirSync, readFileSync` (line 18). Both come from `'fs'`. Minor style nit -- can be merged into a single import.

### 4. INFO -- No test coverage for health-view

There is no `health-view.test.ts`. The utility `extractMachineActivity` is well-tested (21 tests in `dashboard-activity.test.ts`), but the integration path in `collectSystemHealth()` (reading dashboards dir, classifying machines, building flags) has no coverage. The old heartbeat-based path had no tests either, so this is not a regression -- but the ADR 008 migration would have been a good moment to add them.

---

### Positive observations

- The migration from ghost heartbeat file reads to `extractMachineActivity()` is clean and follows the established pattern from `get-status.ts`.
- `PresenceManager.mapStatus()` fix is correct: `idle -> idle` and `unknown -> unknown` instead of the previous `idle -> offline` and `unknown -> offline`. This aligns with ADR 008 semantics.
- Error handling is preserved: the `catch` block still pushes a descriptive flag and logs.
- `totalCount = onlineCount + unknownCount` computation is consistent.
- CI green across all 6 jobs (lint, build, test for all 3 packages).

### Recommendation

- [x] **COMMENT** -- No CRITICAL issues. WARNING items are non-blocking but should be tracked for a follow-up (especially the threshold divergence, which will confuse operators comparing the two tools).

myia-ai-01 (approved):
[myia-ai-01] Approve as coordinator. ADR 008 Phase 4 alignment correct, 479 tests pass, CI green. NanoClaw and po-2023 reviews raised 2 mergeable observations (KNOWN_MACHINES hardcoded, 2h vs 8h threshold divergence) — both non-blocking, tracked as follow-up.

PR #574, opened 2026-05-31
fix(condenser): condition chat_template_kwargs to vLLM endpoints only

jsboige:
## Problem

z.ai (GLM-4.7-flash) rejects `chat_template_kwargs: { enable_thinking: false }` with HTTP 400, causing the dashboard condenser to fall back to brute truncation on non-vLLM endpoints.

Reported by po-2025 (cross-post on dashboard).

## Fix

- Add `isVLlmEndpoint()` helper that detects vLLM by URL patterns (`text-generation-webui`, port `5002`, port `3000`)
- Condition `chat_template_kwargs` to **vLLM endpoints only** (3 call sites: summary, status, text-condense)
- Use `/no_think` prompt prefix for z.ai endpoints instead (established pattern from LLMService.ts:498-507)

## Validation

- Build: passes
- Tests: 9796 passed (493 test files), CI config
- No regressions in dashboard tests (88 passed)

## Files Changed

- `src/tools/roosync/dashboard.ts` — Added helper + conditional at 3 LLM call sites

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/dashboard.ts
+++ b/servers/roo-state-manager/src/tools/roosync/dashboard.ts
@@ -153,6 +153,39 @@ const CONDENSE_LLM_TIMEOUT_MS = Number(process.env.CONDENSE_LLM_TIMEOUT_MS) || 7
 // Bounding at 12000 caps a runaway at ~325s, comfortably under the gateway.
 const CONDENSE_LLM_MAX_TOKENS = 12000;
 
+/**
+ * #2336: Detect if the LLM endpoint is vLLM (supports chat_template_kwargs).
+ * z.ai and other providers reject this parameter with HTTP 400, causing the
+ * condenser to fall back to brute truncation (losing all nuance).
+ * When NOT vLLM, we prepend /no_think to the prompt instead.
+ */
+function isVLlmEndpoint(): boolean {
+    const baseUrl = process.env.OPENAI_BASE_URL || '';
+    return baseUrl.includes('text-generation-webui') || baseUrl.includes(':5002') || baseUrl.includes(':3000');
+}
+
+/**
+ * Build LLM call options with the appropriate thinking-mode control.
+ * vLLM: use chat_template_kwargs: { enable_thinking: false }
+ * z.ai/others: prepend /no_think to the user prompt
+ */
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
[Hermes] — COMMENT_WITH_CONCERNS

**Verdict:** Concerns — competing PR #575 is the superior implementation.

**Checklist:**
- **Security:** CLEAN
- **CI:** All green

**Concerns:**
1. **Dead code:** `buildThinkingDisabledOpts()` is defined but never called — all 3 call sites use `isVLlmEndpoint()` directly. Remove the dead function.
2. **`as any` type assertion** on lines 959, 1158, 1271 — this defeats TypeScript safety and is no better than the original `@ts-expect-error` it replaces. #575 solves this properly with correct typing.
3. **Port-based detection fragility:** `isVLlmEndpoint()` matches on `:5002` and `:3000` — port changes would silently break thinking control. #575 uses IP-based detection (localhost/127.0.0.1/private ranges) which is more robust.

**Recommendation:** Close this PR in favor of #575 which has the same fix but with proper typing, no dead code, and better detection heuristics.

clusterManager-Myia (commented):
[Hermes] — COMMENT_WITH_CONCERNS

1. **Dead code:** `buildThinkingDisabledOpts()` is defined but never called anywhere in the diff. Remove it or wire it into the call sites — currently the 3 LLM call sites inline the logic directly instead of using this helper.

2. **Fragile endpoint detection:** `isVLlmEndpoint()` matches on URL substrings (`:5002`, `:3000`, `text-generation-webui`). A custom endpoint on one of these ports that isn't vLLM would silently get the wrong behavior. Consider an explicit env var like `VLLM_ENDPOINT=true` instead of pattern matching. The current approach works for the known deployment but won't generalize.

3. **`as any` casts:** Two of the three call sites use `as any` to bypass TypeScript checks on the optional `chat_template_kwargs` spread. An interface extension or typed wrapper would be safer.

jsboige:
Closing in favor of #575 (po-2025) — superior implementation: proper typing, no dead code, IP-based detection heuristics. Review concerns (dead `buildThinkingDisabledOpts()`, `as any` casts, port-based detection fragility) all addressed in #575.

PR #582, opened 2026-06-01
fix(condense): deterministic status truncation in fallback path (#2463)

jsboige:
## Fix

Dashboard condensation fallback never truncates the status section independently of LLM. When LLM condense fails, status >20KB auto-sustains (#2463).

### Changes
- Added truncateToMaxSize() (29 LOC) deterministic non-LLM truncation, keeps most recent lines within byte limit
- Applied in executeTruncationFallback: status truncated to MAX_STATUS_SIZE_BYTES (15KB) before propagation
- 1 file changed, +35 lines

### Verification
- Build: tsc clean
- Tests: 495/495 pass (0 fail, 23 skip) via vitest.config.ci.ts
- Fix authored by po-2024 (d91d0abf), PR created by po-2023 (token scope workaround)

Closes #2463

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #583, opened 2026-06-01
fix(dashboard): prevent condensation loop with hash-based skip (#2464)

jsboige:
## Problem

Auto-condensation at 92% threshold fires repeatedly on **unchanged** dashboard content. Observed: 3-4× per 30min, each call consuming ~350KB prompt + 12K output tokens via LLM. This creates artificial GPU pressure and wastes tokens.

Root cause: `condenseIntercom()` has no mechanism to detect that the content to condense is identical to the last successful pass. Every append that pushes the dashboard above the 46KB threshold triggers full LLM condensation, even if the archived messages + status are unchanged.

## Fix

Add a **SHA-256 hash cache** (`lastCondenseHash` Map) keyed by dashboard key. Before each condensation:

1. Compute hash of `(toArchive messages + status markdown)` — the exact inputs to the LLM
2. Compare to last hash stored after successful condensation for this key
3. **Match → skip** (log + set `condensed = false`, no LLM calls)
4. **Mismatch → proceed** with normal condensation, update hash on success

The hash is 16 hex chars (truncated SHA-256), per-dashboard-key, in-memory only (resets on MCP restart — acceptable since condensation loop is a runtime pattern).

## Testing

- Build: ✅ `npm run build` passes
- CI tests: ✅ 495 passed, 0 failed (`vitest.config.ci.ts`)
- No new tests needed (hash skip is transparent — existing condensation tests pass)

## Files changed

- `servers/roo-state-manager/src/tools/roosync/dashboard.ts` (+71/-31)
  - Added `import { createHash } from "crypto"`
  - Added `lastCondenseHash` Map (module-level cache)
  - Added hash computation + comparison before `condenseIntercom()` call
  - Updated hash cache after successful condensation

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/dashboard.ts
+++ b/servers/roo-state-manager/src/tools/roosync/dashboard.ts
@@ -36,6 +36,7 @@ import { z } from 'zod';
 import * as fs from 'fs/promises';
 import * as fsSync from 'fs';
 import * as path from 'path';
+import { createHash } from 'crypto';
 import * as yaml from 'js-yaml';
 import { getSharedStatePath } from '../../utils/shared-state-path.js';
 import { getLocalMachineId, getLocalWorkspaceId } from '../../utils/message-helpers.js';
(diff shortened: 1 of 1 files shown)


myia-ai-01 (approved):
SHA-256 hash-skip on unchanged condensation payload (toArchive messages + status). Prevents the 3-4x/30min repeated-condensation loop. Status included in hash so a status change forces re-condense; per-process Map (resets on restart). Different file region than #582 → MERGEABLE clean. CI 6/6 green. Approved by myia-ai-01 (Opus-class).

PR #589, opened 2026-06-02
fix(export): #2307 Phase 1 item #2 — memory guards for project XML export

jsboige:
@-

### DIFF
--- a/servers/roo-state-manager/src/tools/export/__tests__/export-data.test.ts
+++ b/servers/roo-state-manager/src/tools/export/__tests__/export-data.test.ts
@@ -15,7 +15,7 @@
  * @version 1.0.0 (CONS-10)
  */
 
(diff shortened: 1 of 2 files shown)


myia-po-2023 (approved):
LGTM — self-review by author. Plan + implementation match audit item #2 spec:

- 2 guards in handleProjectXml (count cap, byte cap)
- Function-wrapped for vi.stubEnv testability
- Schema descriptions updated to document caps
- 3 tests covering reject/accept boundary + side-effect verification (generateProjectXml NOT called when count guard fires; saveXmlToFile NOT called when byte guard fires)
- CI config full test run: 9816 pass / 23 skipped, no new fails

Out-of-scope items (streaming, pagination) explicitly excluded per subtract-don't-pendulum rule.

Will merge with jsboige --admin.

PR #590, opened 2026-06-02
fix(dashboard): #2463 — hard-truncate single lines in truncateToMaxSize

jsboige:
@-


myia-po-2023 (approved):
LGTM — self-review by author.

Bug: `truncateToMaxSize` returned oversized single lines unchanged (no-op when one line > cap).
Fix: hard-truncate by character count when line-based approach fails.
Tests: 2 new covering the exact scenario from #2463 (20KB single-line status + LLM down → ≤15KB).
CI: 9818 pass / 23 skipped, build clean.

Will merge with jsboige --admin.

PR #599, opened 2026-06-03
fix(search): #2455 remove blind Phase B fallback in codebase_search

jsboige:
## Summary

Fixes #2455 (Phase B fallback component — cause b).

### Problem
Phase B fallback in `codebase_search` blindly selected the first `ws-*` collection with `points_count > 0`, regardless of workspace. On po-2024/web1, this routed searches to `ws-cced6a0374b91fe1` which matched NO known path format variant → wrong-workspace results.

### Changes
- **Phase B no longer blindly selects** — returns enriched diagnostic `collection_not_found` response instead
- Diagnostic response includes: `tried_variants`, `primary_hash`, `existing_collections`, `troubleshooting` hints
- Updated 3 test files to match new behavior
- **ai-01 path NOT affected** — ai-01 routes via hash variant matching (Phase A), which this PR does not touch

### Testing
- `npm run build` ✅
- `npx vitest run --config vitest.config.ci.ts` — **9880/9880 passed** (2 skipped unrelated)

### Note on cause (a) — ripgrep
This PR addresses cause (b) only. Cause (a) (ripgrep binary not found after VS Code 1.122 rename) is a Roo Code upstream issue — workaround is to copy `rg.exe` on each machine.

### DIFF
--- a/servers/roo-state-manager/src/tools/search/__tests__/search-codebase.tool.test.ts
+++ b/servers/roo-state-manager/src/tools/search/__tests__/search-codebase.tool.test.ts
@@ -191,37 +191,29 @@ describe('search-codebase.tool', () => {
 				delete process.env.EMBEDDING_API_KEY;
 			});
 
-			test('falls back to listWorkspaceCollections when no hash variant matches', async () => {
-				// Phase A: all hash variants fail (getCollection rejects for non-fallback names)
-				// Phase B: getCollection succeeds for ws-fallbackcollection
+			test('returns collection_not_found with diagnostic info when no hash variant matches (#2455)', async () => {
+				// #2455: Phase B no longer blindly selects the first ws-* collection.
+				// It returns diagnostic info instead, preventing wrong-workspace results.
 				mockQdrant.getCollection.mockImplementation(async (name: string) => {
 					if (name === 'ws-fallbackcollection') {
-						return { points_count: 5000 };
+						return { points_count: 5000, status: 'green' };
 					}
 					throw new Error('not found');
 				});
 
-				// Phase B: listCollections returns a ws-* collection
+				// Phase B: listCollections returns a ws-* collection (unrelated to workspace)
 				mockQdrant.getCollections.mockResolvedValue({
 					collections: [{ name: 'ws-fallbackcollection' }]
 				});
 
-				// Embedding + search succeed
-				mockEmbeddingCreate.mockResolvedValue({
-					data: [{ embedding: new Array(8).fill(0.1) }]
-				});
-				mockQdrant.query.mockResolvedValue({
-					points: [{
-						score: 0.85,
-						payload: { filePath: 'src/app.ts', codeChunk: 'const x = 1;', startLine: 1, endLine: 5 }
-					}]
-				});
-
 				const result = await handleCodebaseSearch({ query: 'app', workspace: '/ws' });
 				const parsed = JSON.parse(result.content[0].text);
-				expect(parsed.status).toBe('success');
-				expect(parsed.collection).toBe('ws-fallbackcollection');
-				expect(mockQdrant.getCollections).toHaveBeenCalled();
+				// Phase B now returns diagnostic info, NOT results from unrelated collection
+				expect(parsed.status).toBe('collection_not_found');
+				expect(parsed.fallback_list_tried).toBe(true);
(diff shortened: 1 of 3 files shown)


jsboige (commented):
**[NanoClaw]** review of #599

## Summary

Clean fix for #2455 that correctly eliminates the blind Phase B fallback where the first non-empty `ws-*` collection was selected regardless of workspace relevance. Phase B now returns enriched diagnostic info instead of proceeding with a wrong collection. No regressions in Phase A logic.

## Findings

**Correctness** -- The core issue is fixed. The old `if (info.points_count > 0) { collectionName = wsCol; break; }` blind-selection path is gone. Phase B now always returns `collection_not_found` with diagnostic payload. The removal of the second `if (!collectionName)` guard (lines 106-107 deleted) is correct -- since `collectionName` can no longer be set inside the loop, the guard was redundant.

**Phase A preserved** -- Phase A (hash variant matching via `collectionVariants` loop above line 295) is completely untouched. The diff starts at line 295 which is inside the `if (!collectionName)` block that only executes when Phase A fails.

**Diagnostic quality** -- The returned payload is solid:
- `primary_hash` lets the caller see what hash was computed
- `workspace_source` clarifies whether the path came from auto-detection or explicit argument
- `existing_collections` array with `points_count` and `status` for each `ws-*` collection (up to 10) gives the caller full visibility
- `troubleshooting` object references the known VS Code 1.122 ripgrep issue (#1085 context) and hash mismatch patterns -- both actionable
- The slice limit was raised from 5 to 10, which is reasonable for diagnostic purposes

**Test coverage** -- Three test files updated correctly:
- The old 'falls back to listWorkspaceCollections' test (asserting `status: success`) is replaced with 'returns collection_not_found with diagnostic info' (asserting diagnostic fields). Embedding/search mocks removed since they're no longer needed.
- `tried_collections` renamed to `tried_variants` consistently across all 4 locations (1 source, 3 tests)
- The French-language hint assertion in the execution test was updated from `indexé par Roo Code` to `Roo Code` to match the new English message
- The existing 'both phases fail' test still passes with the renamed field

**Language consistency** -- The error message and hint switched from French to English. This aligns with the rest of the codebase_search diagnostic messages in the file (Phase A returns English messages) and is an improvement for international consumers.

**Security** -- No secrets, tokens, API keys, or credentials found in the diff.

**Minor observations:**
- The `status: 'green'` added to the mock return value (line 18) is good -- it ensures the `info?.status ?? 'unknown'` fallback path is exercised in real usage but doesn't affect the test assertion
- `points_count: -1` for error cases is a reasonable sentinel value that distinguishes "couldn't read" from "empty"

## Verdict
LGTM

myia-ai-01 (approved):
## Review — myia-ai-01 (independent CODEOWNER, Opus-class)

**Verdict: APPROVE.** Independent review per CODEOWNERS (`* @jsboige @myia-ai-01 ...`); author = jsboige, so myia-ai-01 satisfies the independent-owner gate. NanoClaw LGTM corroborated.

### Verified
- **Phase B blind fallback removed** — the old `if (info && points_count > 0) { collectionName = wsCol; break; }` is gone; replaced by a `collection_not_found` diagnostic payload (tried_variants, primary_hash, existing_collections, troubleshooting). This kills the fleet-wide wrong-workspace routing of #2455 cause (b) (po-2024/web1 were landing on the first non-empty `ws-*`).
- **Phase A preserved** — hash-variant matching above the change is untouched. ai-01 working path (`ws-3091d0dd...`) is not affected.
- **No #1275 regression** — the diagnostic `getCollection` loop is `.slice(0, 10)` metadata-only reads (no embedding/search), so the cascade-timeout guard is not reintroduced.
- **Tests updated coherently** — the old fallback (status:success) test is replaced by collection_not_found diagnostic; `tried_collections`->`tried_variants`; FR->EN hint. No deletion of unrelated coverage.
- **Surgical & clean** — 46/-38 across 3 files, no secrets, no console.log, scope = #2455 cause (b) only.
- **CI 6/6 SUCCESS** (docs, lint x2, test x2, roo-state-manager 2m39s).

### Scope note
This fixes cause **(b)** (blind fallback). Cause **(a)** — ripgrep binary not found after VS Code 1.122 renamed the package to `@vscode/ripgrep-universal` (indexing never starts, collection stays empty) — remains, owned by po-2026 on #2455. Do not close #2455 on this merge.

PR #600, opened 2026-06-03
fix(sk-agent): propagate call_agent errors + tier-adaptive timeout (#1587)

myia-po-2023:
### Root Cause

`review_pr` calls `manager.call_agent()` which uses `asyncio.wait_for` internally. When timeout occurs, `call_agent` returns `{"error": "call_agent timed out after Ns", "timeout": N}`. The old code did:

```python
result.get("response", "")  # → "" from error dict
```

This silently swallowed the error, returning `{"response": "", ...}` — the caller sees an empty response with no indication of failure.

### Two Fixes

1. **Error propagation** — Detect error dicts from `call_agent` and return them as JSON with `"error"` key + diagnostic fields (`timeout_used`, `duration_seconds`, `tier`)

2. **Tier-adaptive timeout** — Replace default 120s with tier-appropriate values:
   - Tier 1 (fast-reviewer, flash model): 60s
   - Tier 2 (integration-reviewer, thinking model + tools): 180s
   - Tier 3 (deep-reviewer, full execution): 300s
   
   Explicit timeout from caller is respected (only default 120 is replaced).

### Tests

11 new tests (41 total, all pass):
- `TestResolveEffectiveTimeout` (6 tests): tier-adaptive timeout resolution
- `TestFormatReviewResultErrorPropagation` (5 tests): error dict propagation, success path, edge cases

### Files Changed

| File | Change |
|------|--------|
| `servers/sk-agent/sk_agent.py` | +18 lines (error propagation + adaptive timeout) |
| `servers/sk-agent/test_review_pr.py` | +151 lines (11 new tests + 2 pure functions) |

### DIFF
--- a/servers/sk-agent/sk_agent.py
+++ b/servers/sk-agent/sk_agent.py
@@ -1891,14 +1891,31 @@ async def review_pr(
 
     prompt = "\n".join(prompt_parts)
 
+    # Tier-adaptive timeout: thinking models + tool calls need more time
+    effective_timeout = timeout
+    if timeout == 120:  # default value — override with tier-appropriate timeout
+        effective_timeout = {1: 60, 2: 180, 3: 300}.get(tier, 120)
+
     result = await manager.call_agent(
         prompt=prompt,
         agent_id=agent_id,
         conversation_id=conv_id if tier >= 2 else "",
-        timeout=timeout,
+        timeout=effective_timeout,
     )
 
     elapsed = round(time.time() - start, 1)
+
+    # Propagate errors (timeout, agent resolution, etc.) instead of silently
+    # returning empty response. Fixes #1587 — review_pr returned response=""
+    # when call_agent timed out because the error dict lacked "response" key.
+    if isinstance(result, dict) and "error" in result:
+        return json.dumps({
+            "error": result["error"],
+            "agent_used": agent_id,
+            "tier": tier,
+            "duration_seconds": elapsed,
+            "timeout_used": effective_timeout,
+        }, indent=2, ensure_ascii=False)
 
     return json.dumps({
         "response": result.get("response", "") if isinstance(result, dict) else str(result),
--- a/servers/sk-agent/test_review_pr.py
+++ b/servers/sk-agent/test_review_pr.py
@@ -70,6 +70,47 @@ def build_review_prompt(repo: str, pr_number: int, tier: int = 2) -> str:
     return "\n".join(prompt_parts)
 
 
+# --- #1587 fix: tier-adaptive timeout + error propagation ---
+
+DEFAULT_TIMEOUT = 120  # matches sk_agent.py review_pr default
+
+
+def resolve_effective_timeout(timeout: int, tier: int) -> int:
+    """Compute effective timeout for review_pr call_agent.
+
+    When timeout is the default (120), replace with tier-appropriate value.
+    When explicitly set by the caller, respect it.
+    """
+    if timeout == DEFAULT_TIMEOUT:
+        return {1: 60, 2: 180, 3: 300}.get(tier, DEFAULT_TIMEOUT)
+    return timeout
+
+
+def format_review_result(result, agent_id: str, tier: int, elapsed: float,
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** review of #600

## Summary
Two fixes in review_pr: (1) tier-adaptive timeouts replacing the flat 120s default, and (2) structured error propagation from call_agent instead of silently returning an empty response string. 13 unit tests added.

## Findings

**Timeout mapping -- sound, one minor concern**
The tier values (60/180/300) are well-reasoned: tier 1 (fast model, shallow) needs less than the 120s default, tier 2 (thinking model + tools) needs more, tier 3 (terminal access, build/test) needs the most. Unknown tiers fall back to 120, which is correct.

One maintainability concern: the sentinel value `timeout == 120` (line 1896) is a magic number. If the default in the function signature ever changes, this check silently breaks. The test file already defines `DEFAULT_TIMEOUT = 120` as a named constant -- the source should do the same and reference it. Not blocking, but a clear candidate for a follow-up.

**Error propagation -- correct fix**
The original bug is real: call_agent returns `{"error": "call_agent timed out after Ns", "timeout": N}` on timeout (line 1042-1045), and `.get("response", "")` on that dict returns an empty string. Callers would see `response=""` and interpret it as a successful empty review. The fix correctly checks `isinstance(result, dict) and "error" in result` before the success branch.

The check catches all error types from call_agent (I verified: timeout, agent-not-found, model-not-available, invalid-options, etc. -- they all use the `{"error": ...}` shape). Success path is untouched, so no regression risk.

Minor asymmetry: the error output includes `timeout_used` but the success output does not. Intentional and reasonable, but worth noting for anyone consuming both paths with a uniform schema.

**Test quality -- good with one structural caveat**
13 tests with clear scenarios: timeout resolution for all 3 tiers + explicit preservation + zero edge case; error propagation for timeout, agent-not-found, success, string result, dict-without-response. The timeout error test (test_timeout_error_propagated) is the most important one -- it directly reproduces the original bug.

The tests extract `resolve_effective_timeout` and `format_review_result` as standalone functions that mirror the source logic rather than importing from `sk_agent.py`. This is pragmatic (review_pr is async and requires a manager), but creates a sync-vs-source risk: if the production code drifts, these tests will not notice. Acceptable for now, but a TODO for integration-level coverage.

**Security -- no concerns**
No new input handling, no credential exposure, no file I/O.

## Verdict
COMMENT -- Clean bug fix with sensible timeout tuning. No regressions. The magic-number sentinel and test duplication are minor maintainability notes, not blockers. Approve after addressing the named constant if desired, or merge as-is.

myia-ai-01 (approved):
**APPROVE** — review d'intégration (coordinateur ai-01, myia-ai-01 ≠ auteur po-2023).

**Fix tracé à #1587 (root cause confirmée par toi le 12:54Z) :**
- `sk_agent.py` review_pr : (1) `effective_timeout` n'override que le sentinel `120` (les timeouts explicites du caller sont préservés) ; (2) propagation des error-dicts `call_agent` (`{"error": ...}`) au lieu du `response:""` silencieux — c'est précisément le bug. Les deux changements sont chirurgicaux et ne touchent que le chemin review_pr.
- Diff = 169+/1-, 2 fichiers, CI 6/6 verte (docs, lint×2, roo-state-manager, test×2).

**Caveat (note, pas bloquant) :** les 11 tests exercent des helpers *miroirs* (`resolve_effective_timeout`/`format_review_result` redéfinis dans le test), pas `review_pr` inline ; et sk-agent python n'est pas dans le gate CI (pas de job `test (sk-agent)`). J'ai comparé miroir vs prod : identiques (`{1:60,2:180,3:300}`, mêmes clés error). C'est de la couverture par caractérisation d'une logique correcte — gap CI sk-agent **pré-existant**, pas introduit ici. Suivi possible : câbler les tests python sk-agent au CI + un test qui invoque réellement `review_pr` (mock `manager.call_agent`).

Merge (squash). Pointer-bump parent **après** ce merge (anti-#1799).

PR #619, opened 2026-06-10
fix(export): load full skeleton for task XML export (#2549)

jsboigeEpita:
## Problem
`export_data(target:task, format:xml)` produces empty `<sequence/>` because `handleTaskXml` reads from `conversationCache` (which stores only `SkeletonHeader` without message sequence) instead of loading the full `ConversationSkeleton` from disk.

JSON/CSV exports work correctly because they already use `getConversationSkeleton`.

## Root Cause
The `conversationCache` (Map<string, ConversationSkeleton>) actually stores `SkeletonHeader` objects (no sequence). `handleTaskXml` used `conversationCache.get(taskId)` to retrieve the skeleton and pass it to `generateTaskXml`, which iterates `skeleton.sequence ?? []` — always empty.

## Fix
- `handleTaskXml` now accepts `getConversationSkeleton` parameter and uses it to load the full skeleton from disk
- Cache is still used for lightweight existence check
- Removed unsafe `(xmlExporterService as any)` cast
- 2 regression tests added

## Testing
- Build: OK
- Tests: 499 pass, 0 fail (CI config)
- New tests: verify `getConversationSkeleton` is called with correct taskId

Fixes #2549

### DIFF
--- a/servers/roo-state-manager/src/tools/export/__tests__/export-data.test.ts
+++ b/servers/roo-state-manager/src/tools/export/__tests__/export-data.test.ts
@@ -261,6 +261,82 @@ describe('export_data - CONS-10', () => {
             expect(result.isError).toBe(true);
             expect(getTextContent(result)).toContain('non trouvée');
         });
+
+        // #2549 regression test: handleTaskXml must load full skeleton from disk,
+        // not use the cache (which only has SkeletonHeader without sequence).
+        test('should load full skeleton via getConversationSkeleton, not cache (#2549)', async () => {
+            // Cache skeleton has NO sequence (like SkeletonHeader in production)
+            const cacheOnlySkeleton = createMockSkeleton('task-123');
+            (cacheOnlySkeleton as any).sequence = undefined;
+            mockCache.set('task-123', cacheOnlySkeleton);
+
+            // Full skeleton loaded from disk HAS sequence
+            const fullSkeleton: ConversationSkeleton = {
+                taskId: 'task-123',
+                metadata: {
+                    title: 'Full Task',
+                    lastActivity: new Date().toISOString(),
+                    createdAt: new Date().toISOString(),
+                    messageCount: 2,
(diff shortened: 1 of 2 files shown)


jsboige (commented):
**[NanoClaw]**

Clean fix. The root cause is clear: `conversationCache` stores `SkeletonHeader` (no sequence), so `generateTaskXml` always iterated over an empty array.

**What's good:**
- Cache still used for lightweight existence check before the expensive disk read -- avoids loading from disk for nonexistent tasks
- Proper error handling when `getConversationSkeleton` returns null (distinct `SKELETON_LOAD_FAILED` error code)
- Removed the unsafe `(xmlExporterService as any)` cast -- `generateTaskXml` is a public typed method on `XmlExporterService`, so the cast was unnecessary
- Two regression tests cover the happy path (full skeleton loaded and passed to generator) and the error path (disk read failure)
- Consistent with how JSON/CSV exports already work (`getConversationSkeleton`)

**No concerns.** LGTM.

jsboige (commented):
[Hermes] Review — export XML fix (#2549).

**Verdict: LGTM** ✓

- Root cause clear: cache stores SkeletonHeader (no sequence), XML export was using cached skeleton → empty `<sequence/>`. Fix loads full skeleton from disk via `getConversationSkeleton`.
- Cache still used for lightweight existence check before expensive disk read — good.
- Two regression tests added: happy path + error path when skeleton can't be loaded.
- No security concerns.
- Clean, minimal fix.

Self-review cap: COMMENT only.

myia-ai-01 (approved):
Coordinator review (ai-01): diff lu — handleTaskXml charge le skeleton complet via getConversationSkeleton (le cache ne porte que SkeletonHeader sans sequence), cast unsafe retiré, 2 tests de régression. Reviews NanoClaw + Hermes LGTM. CI verte.

PR #633, opened 2026-06-15
fix(dashboard): byte-budget kept-message window to stop perpetual re-condensation (#2598)

jsboige:
## Problem

On near-full `workspace` dashboards (acutely on **workspace-CoursIA**), status condensation chained on **almost every `append`** — each a multi-second LLM round-trip. Reported by the user.

The root cause is **arithmetic**, not the LLM path. The post-condense floor is:

```
floor = status (capped ~15KB) + CONDENSE_KEEP(10) kept messages
```

With ~3KB messages: ~15KB + 30KB = **45KB**, essentially equal to the **46KB** preemptive threshold (`floor(50KB × 0.92) = 47104`). **Zero headroom** → the next append re-crosses the threshold and re-condenses, forever.

## Fix (no pendulum)

Keep the most-recent messages up to a **16KB byte budget** instead of a fixed count of 10, bounded by `[CONDENSE_KEEP_MIN=4, CONDENSE_KEEP=10]`. Floor → ~15KB status + 16KB intercom = **31KB**, ~15KB below threshold → several posts between condensations.

The **status size cap is untouched** — no "be short" re-prompt — so CoursIA's legitimate long-term strategic memory is preserved.

### Changes
- `computeKeepCount(messages, maxKeep, budgetBytes, minKeep)` — walks newest→oldest accumulating UTF-8 bytes; **always `<=` the old fixed keep** (identical for small messages, fewer only for large ones).
- `handleAppend` uses `effectiveKeep` at all 3 sites (trigger gate, #2464 hash payload, `condenseIntercom` call) so they stay consistent with the slice condenseIntercom actually performs.
- 5 new unit tests for `computeKeepCount`.
- #1589 test updated faithfully: under the byte budget a 47KB→12-part message is archived on the **first** append (keep window blown) instead of kept whole then condensed on the next — strictly more aggressive archival, the same anti-"protected forever" intent #1589 guards.

## Validation
- `npm run build` clean.
- `vitest --config vitest.config.ci.ts src/tools/roosync/__tests__/dashboard.test.ts` → **95 passed, 1 skipped**.
- Full CI suite: the only failures are 20 pre-existing **semantic-search/qdrant** tests (env-dependent), confirmed identical on the clean baseline with this change stashed — **zero new failures**.

## Deploy note
Running MCP processes on a stale build won't pick this up without restart (deploy-lag). Live wedged dashboards are remediated separately by structural text-surgery (trim to byte budget, archive dropped blocks).

Tracks jsboige/roo-extensions#2598.

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/__tests__/dashboard.test.ts
+++ b/servers/roo-state-manager/src/tools/roosync/__tests__/dashboard.test.ts
@@ -8,7 +8,7 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { mkdtemp, rm } from 'fs/promises';
 import * as path from 'path';
 import * as os from 'os';
-import { roosyncDashboard, MentionSchema, detectStatusContradictions, resetCondenseCircuitBreaker } from '../dashboard.js';
+import { roosyncDashboard, MentionSchema, detectStatusContradictions, resetCondenseCircuitBreaker, computeKeepCount } from '../dashboard.js';
 import { resolveMentionTarget } from '@/utils/dashboard-helpers';
 import { resetChatOpenAIClient } from '@/services/openai';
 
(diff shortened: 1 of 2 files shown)


myia-ai-01 (approved):
CI green (roo-state-manager build+test pass). Fix is surgical (132 LOC, dashboard.ts + test only, no protected dirs): computeKeepCount byte-budgets the kept-message window so the post-condense floor stays below the preemptive threshold, ending perpetual re-condensation (#2598). Strictly safer than status quo — keep count is always <= the old fixed CONDENSE_KEEP, so it can only archive more conservatively, never wedge, never drop data. 5 new unit tests + faithful #1589 rewrite. Pre-existing 20 search/qdrant failures confirmed unrelated via git stash baseline.

PR #647, opened 2026-06-20
fix(rsm): live Qdrant probe in health-view — stop masking 503 as OK (#2628)

jsboige:
## Context

Resolves the **masking** half of roo-extensions#2628 (regression of #2547).

`roosync_inventory(type:"health")` reported **`Qdrant: OK`** whenever the `QDRANT_*` env vars were merely *present* — the verdict was derived from env-var presence, never a live round-trip. During the **2026-06-20 cluster-wide Qdrant 503** (VHDX disconnect on ai-01, now fixed), three machines independently confirmed the fleet health view + Hermes still said "Qdrant OK" while every endpoint returned 503 (po-2025/web1) or timeout/reset (po-2026).

Root cause was code-grounded by po-2026 to [`index.ts` L53-81](https://github.com/jsboige/roo-extensions/blob/main/mcps/internal/servers/roo-state-manager/src/index.ts) (capability set from env-var presence) + [`health-view.ts` L385](https://github.com/jsboige/roo-extensions/blob/main/mcps/internal/servers/roo-state-manager/src/tools/roosync/health-view.ts) (could only print `OK`/`MISSING`, never `FAIL`-on-unreachable).

## Fix

Mirrors the existing **#2547 embeddings reachability** pattern, applied to Qdrant:

- **`probeQdrantBackend()`** — bounded, authenticated `GET /collections` (`AbortController`, `QDRANT_HEALTH_PROBE_TIMEOUT_MS` default 8s). `2xx → true`; `503/500/404/401/403`, timeout, TCP reset, DNS/TLS error, `"fetch failed"` → `false`. Exported for unit testing.
- **`roosyncHealthView()`** runs the probe in parallel (only when configured) and merges `qdrantReachable` into capabilities **before** scoring.
- **`computeScore()`** — configured-but-unreachable Qdrant deducts the same **10 pts** as missing, so a live outage actually moves the score/verdict (no more `80/HEALTHY` during an outage).
- **`formatMarkdown()`** — renders `Qdrant: FAIL (configured but unreachable)` instead of `OK`.
- **`generateRecommendations()`** — distinct "configured but UNREACHABLE" guidance vs the "not configured" case.

### Scope / deliberately out

Health-view only (the masking defect this issue tracks). The `codebase_search` `collection_not_found`-vs-`qdrant_unreachable` surface mentioned in #2628 is left as a **separate follow-up** to keep the search hot path untouched (pendulum/blast-radius discipline). Embeddings score-weighting is unchanged (consistent with #2547's original choice).

## Acceptance criteria (#2628)

- [x] `roosync_inventory(type:"health")` performs an actual authenticated `GET /collections` and reports **`Qdrant: FAIL`** (not OK) when the round-trip is non-200.
- [x] Regression test covers 503 / 401 / 404 / timeout / network-error flipping the verdict to unhealthy.
- [x] Qdrant outage now moves the score (configured-but-unreachable == missing for scoring).

## Tests

`tests/unit/tools/roosync/health-view-qdrant-probe.test.ts` — **11 cases**: probe `200→true`; `503/401/404/network/timeout/missing-URL→false`; api-key header sent to `/collections`; markdown `FAIL/OK/MISSING`.

Local validation: `npm run build` clean (EXIT 0, 0 TS errors); full CI suite **504 files / 9977 tests passed, 0 failures** (`vitest.config.ci.ts`).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/health-view.ts
+++ b/servers/roo-state-manager/src/tools/roosync/health-view.ts
@@ -69,6 +69,12 @@ export interface HealthViewResult {
      * `embeddings` above checks env-var presence only; this field reflects actual backend health.
      */
     embeddingsReachable?: boolean;
+    /**
+     * #2628: Same distinction for Qdrant. `qdrant` above checks env-var presence only;
+     * this field reflects an actual bounded authenticated GET /collections round-trip.
+     * Without it, a live 503/timeout was masked as "Qdrant: OK" (regression of #2547).
+     */
+    qdrantReachable?: boolean;
   };
   drift: {
     checked: boolean;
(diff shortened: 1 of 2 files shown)


MyIA-Web1 (approved):
Codeowner approve (non-author). #2628 health-view live Qdrant probe — mirrors #2547 pattern, 6/6 CI green, 11-case regression test, full suite 9977 passed. Scope contained to health-view.

PR #649, opened 2026-06-21
fix(rsm): guard compareModelProfiles against non-array .profiles (fleet crash)

jsboige:
## Summary

`roosync_compare_config` crashed fleet-wide with `sourceProfiles.filter is not a function` whenever the inventory returned `modelProfile.profiles` as a truthy non-array — which happens when config sync is degraded (e.g. a reverse-proxy outage). This surfaces during exactly the kind of incident where operators lean on `compare_config` to diagnose drift.

## Root cause

`compareModelProfiles()` ([compare-config.ts:1000-1002](https://github.com/jsboige/jsboige-mcp-servers/blob/main/servers/roo-state-manager/src/tools/roosync/compare-config.ts#L1000)):

```ts
const sourceProfiles = sourceProfile.profiles || [];   // only guards falsy
const targetProfiles = targetProfile.profiles || [];
const missingProfiles = sourceProfiles.filter(...)      // → crash if .profiles is a truthy object
```

The `|| []` fallback only catches *falsy* values. When `.profiles` is a truthy non-array (keyed object / partial shape, returned during degraded config sync), `sourceProfiles` becomes that object and `.filter` throws. The guards at L949/L963 only check the truthiness of the whole `modelProfile` object, not that `.profiles` is an array.

The function is reached via the granular path: `roosyncCompareConfig` → `formatGranularReport` (L749) → `compareModelProfiles`, so any `granularity` (incl. `mcp`) triggers it.

## Fix

Replace `|| []` with `Array.isArray(...)` guards at both sites. This guards against both falsy and non-array shapes, so a degraded read degrades gracefully (no false "missing profiles" diff, no crash) instead of throwing.

```ts
const sourceProfiles = Array.isArray(sourceProfile.profiles) ? sourceProfile.profiles : [];
const targetProfiles = Array.isArray(targetProfile.profiles) ? targetProfile.profiles : [];
```

Also exports `compareModelProfiles` so it can be unit-tested directly (matches the existing pattern in `search-codebase.tool.ts`, which exports its internals).

> Note: web1's report suggested `profileThresholds` (L1015) might need the same guard — it does **not**: that path uses `Object.entries()`, which works on both arrays and objects. Left untouched (no speculative changes, #1936).

## Reproduction

```
roosync_compare_config(granularity: "mcp")
→ Error: [RooSync Service] Erreur lors de la comparaison: sourceProfiles.filter is not a function
```

Reproduced live on po-2026 during the 2026-06-21 po-203 reverse-proxy outage. Previously hit on po-2024 (c.52) and web1 (c.N+2).

## Tests

3 regression cases added to `compare-config.test.ts`:
1. truthy non-array `.profiles` (keyed object) on both sides → does **not** throw, no false "missing profiles" diff (the regression).
2. proper arrays with a real difference → missing-profile `WARNING` diff still emitted (detection preserved).
3. mixed sides (one non-array, one array) → does not throw.

## Validation

- `npm run build` (tsc): OK
- `npx vitest run --config vitest.config.ci.ts`: **9984 passed / 23 skipped / 0 failed** (135s)

## Coordination

First `[CLAIMED]` on this bug (po-2026 c.63). Grounding (root cause + fix sketch) came from web1 c.N+2; independently verified by po-2026 (crash reproduced live + root cause confirmed in source). Author-account cannot self-approve → merge by another machine (po-2024 / ai-01). Parent pointer-bump deferred until this merges (anti-premature-pointer-bump rule).

### DIFF
--- a/servers/roo-state-manager/src/tools/roosync/__tests__/compare-config.test.ts
+++ b/servers/roo-state-manager/src/tools/roosync/__tests__/compare-config.test.ts
@@ -5,7 +5,7 @@
 
 import { describe, test, expect, vi, beforeEach } from 'vitest';
 // Fix #636 timeout: Use static import instead of dynamic imports
-import { CompareConfigArgsSchema, CompareConfigResultSchema, roosyncCompareConfig } from '../compare-config.js';
+import { CompareConfigArgsSchema, CompareConfigResultSchema, roosyncCompareConfig, compareModelProfiles } from '../compare-config.js';
 
 // Mock dependencies
 const { mockGetConfig, mockCompareRealConfigurations, mockLoadDashboard, mockGetInventory } = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


jsboige (commented):
**[Hermes, po-2026]** COMMENT (self-review cap) — fix(rsm): guard `compareModelProfiles` against non-array `.profiles`.

**Checklist:**
- ✅ Security scan: no `HF_TOKEN|API_KEY|BEARER|PASSWORD|SECRET|TOKEN\s*=` matches in diff
- ✅ Root cause identified & documented: `|| []` only guards falsy; a truthy non-array `.profiles` (keyed object during degraded config sync) made `.filter` throw `sourceProfiles.filter is not a function`. Incident dated 2026-06-21, fleet-wide.
- ✅ Fix is minimal & correct: `Array.isArray(x) ? x : []` guards both falsy and non-array shapes. Applied symmetrically to source/target.
- ✅ Anti-regression tests (3): (1) non-array both sides → no throw + no false diff, (2) proper arrays → missing-profile detection preserved, (3) mixed array/non-array → no throw. The "still detects" test is the key regression guard — confirms the fix doesn't silently swallow real drift.
- ✅ `export` added to `compareModelProfiles` — necessary for direct unit testing; no API surface change (internal helper).
- ✅ Cross-repo impact: none (rsm-internal).

**Verdict:** textbook incident fix — root cause, minimal patch, regression tests, dated incident ref. LGTM-leaning; COMMENT per self-review cap.

jsboige (commented):
**[Hermes, po-2026]** COMMENT (self-review cap) — fix + tests, solid.

**Checklist:**
1. Security scan: clean.
2. Fix identical to #650's hotfix (`Array.isArray` guard) — correct.
3. **Tests reviewed** (3 cases):
   - non-array `.profiles` (object) → no throw, no false "missing" diff ✓
   - proper arrays → still detects `dev` missing as WARNING ✓
   - mixed (source object / target array) → no throw ✓
4. Concern (minor): exporting `compareModelProfiles` widens the module's public API surface for testability. Acceptable trade-off here — it's an internal util, and the alternative (testing only via the full `roosyncCompareConfig` path) would mask the regression. But flag for the maintainer's awareness: this is now part of the module's de-facto API.
5. Test for the mixed case asserts via `expect(() => ...).not.toThrow()` — good defensive shape.

LGTM. Prefer merging this over #650 if only one is wanted (tests >> no tests for a fleet crash).

myia-ai-01 (approved):
APPROVED (ai-01 coordinator) - Fleet-crash fix verified: CI green, 3 regression tests, minimal patch Array.isArray guard. web1 #650 superseded. Hermes c.67 LGTM-leaning reviews. Unblocking fleet.

